PR #4, opened 2019-07-27
Please review Linked List 2

MahiteeS:


DeleteNodeWithoutHead.java
```diff
@@ -0,0 +1,33 @@
+package linkedList2;
+
+/*
+Idea : 1. We shift the pointer of the node to be deleted to the temp node, point temp node to next node,
```
ashmeetdeol19: Thats not a good explanation.
You need  to improve:
We copy data of next node to the given node and delete the next node.
Thats it.

ReorderList.java
```diff
@@ -0,0 +1,59 @@
...
+            }
+            else next = dq.pollLast();
+                
+            prev.next = next;
+            prev = next;
+            count ++;      
+        }
+        prev.next = null;
+    }
+}
```
ashmeetdeol19: Thats good.
But try the optimal solution as well.

IntersectionOfTwoLinkedLists.java
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }
+        while (headA != headB) {
+            headA = headA.next;
+            headB = headB.next;
+        }
+        return headA;
+    }
+	
+}
```
ashmeetdeol19: Thats a good Solution :)

PR #5, opened 2019-07-27
Tried first problem

lavanganji:


Problem3.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+            
+        while(headA!= headB){
+           headA = headA.next ;
+           headB = headB.next;
+        }
+        
+        return headA;
+    }
+}
```
ashmeetdeol19: Well Done Lavan.
Also try the other solution posted in Slack Channel.

Problem1.java
```diff
@@ -0,0 +1,63 @@
...
+    
+    public static void reorderList(ListNode head) {
+        if (head == null) {
+            return;
+        }
+        ListNode slow = head;
+        ListNode fast = head;
+
+        while (fast != null && fast.next != null) {
+            slow = slow.next;
```
ashmeetdeol19: Just check position of slow and fast pointer in odd and even case.

PR #10, opened 2019-07-28
Commit for Linked List 2

Hariae:


IntersectionList.java
```diff
@@ -0,0 +1,82 @@
...
+
+
+
+        return null;
+    }
+
+    public static void main(String[] args){
+        System.out.println("IntersectionList");    
+    }
+}
```
ashmeetdeol19: Thats a good solution 👍 
But also try the solution without calculating length of lists.

PR #13, opened 2019-08-23
Problem 2 and 3

pgoyal0111:


Remove Node without head
```diff
@@ -0,0 +1,35 @@
...
+}
+*/
+// This function should delete node from linked list. The function
+// may assume that node exists in linked list and is not last node
+// node: reference to the node which is to be deleted
+class GfG
+{
+    void deleteNode(Node node)
+    {   
+        while(node.next.next!=null)
```
sehaj-grover: you don't need to modify every element. Try without the loop

PR #22, opened 2019-09-13
LinkedList-2

SwetaK951:
Please review.

### DIFF
--- a/Deletewithoutheadpointer.java
+++ b/Deletewithoutheadpointer.java
@@ -1,0 +1,21 @@
+class Deletewithoutheadpointer
+{
+    // Time complexity O(1)
+    // Space complexity O(1)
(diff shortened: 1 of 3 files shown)


Deletewithoutheadpointer.java
```diff
@@ -0,0 +1,21 @@
+class Deletewithoutheadpointer
+{
+    // Time complexity O(1)
+    // Space complexity O(1)
+    // Tested , # of attemp =1
+    void deleteNode(Node node)
+    {
+        if(node == null)
+            return;
+        else if(node.next == null)
```
sehaj-grover: this case is not relevant

SwetaK951: Yeah, correct. I understood now.

PR #36, opened 2019-11-02
LinkedList-2

pkbaharani:


DeleteNode with head
```diff
@@ -0,0 +1,53 @@
...
+   
+   while(temp.next!=null&&temp.next.next!=null)
+   {
+       temp.data=temp.next.data;
+       temp=temp.next;
+       
+   }
+    temp.data=temp.next.data;
+    temp.next=null;
+    }
```
sehaj-grover: do we need a while loop?

pkbaharani: Yes, because we have to shift all the elements from the given node to the left, so while loop is required according me.

sehaj-grover: Can't we just skip the next node and continue instead of the while loop

pkbaharani: the name of the file I chose is incorrect, it is without a head pointer, so we are given the node that has to be deleted.
for that, the only option we have is to shift all the elements to the left. this would overwrite the node that has to be deleted.

sehaj-grover: Prateek please watch the session video and see if you can implement this problem without the loop

PR #39, opened 2019-11-02
Reorder and intersection

88sanjay:


sehaj-grover (commented):
Sanjay try the third problem as well
And try using meaningful variables

PR #43, opened 2019-11-03
Please review "DeleteNode, ReorderList and IntersectionOf2Lists"

PrathushaKoouri:


51_deleteANode.py
```diff
@@ -0,0 +1,10 @@
+class ListNode(object):
+    def __init__(self, x):
+        self.val = x
+        self.next = None
+
+def deleteNode(curr_node):
+    #code here
+    if curr_node.next:
```
sehaj-grover: why do you need this if

PR #51, opened 2019-12-30
Completed 2 except problem 1

RishikaMachina:


Problem_2.py
```diff
@@ -0,0 +1,7 @@
+# Runs on Geeks for Geeks
+
+
```
neha79091: Please write complexity

PR #52, opened 2020-01-01
submit linkedlist2

samhithaaaa:


intersectionlinkedlist.py
```diff
@@ -0,0 +1,42 @@
+# Definition for singly-linked list.
+# Time Complexity : O(n)
+# Space Complexity : O(1) 
+# Did this code successfully run on Leetcode : Yes
```
neha79091: Time complexity will be O(m+n)

PR #73, opened 2020-02-28
Intersection of lists - initial update

saiswaroopreddy:


Lohith-Aswathappa:
Good job Sai!

Submit the other problems as well

PR #74, opened 2020-02-29
Solution - Linked-List-2

saumilsjsu:


solution.md
```diff
@@ -0,0 +1,75 @@
+# Problem 1
+## Time Complexity :
+O(n)
```
BhushanVasisht: O(m+n)

BhushanVasisht:
The problem was in reverse function and finding mid point

solution.md
```diff
@@ -43,33 +43,75 @@ No.
...
+                    slow=curr
+
+
+            def _reverse(self,head):
+                curr=head
+                prev=None
+                fast=head.next
+                while(fast!=None):
+                    curr.next=prev
+                    prev=curr
```
super30admin: `curr = fast ` is missing here so list in not getting reversed properly

solution.md
```diff
@@ -43,33 +43,75 @@ No.
...
+                    fast=fast.next
```
super30admin: curr = fast is missing in reverse list

PR #82, opened 2020-03-01
Completed LinkedList2

surajrv35:


super30admin:
Good to see 1 pass solution of intersection lists. Keep working hard

PR #135, opened 2020-06-20
Completed Linked-List-2 : Please review.

shantanu609:


AbhiniveshP:
Nice work on all the 4 questions Shantanu!!!
For questions related to Binary Trees, if you use recursion or a custom stack, mostly the time complexities would be varying with Height of the Tree. So, it would be precise if you mention Time Complexity as O(H) and let know that would end up being O(N) in the worst case.

AbhiniveshP:
Nice work on all the 4 questions Shantanu!!!
For questions related to Binary Trees, if you use recursion or a custom stack, mostly the time complexities would be varying with Height of the Tree. So, it would be precise if you mention Time Complexity as O(H) and let know that would end up being O(N) in the worst case.

PR #172, opened 2020-08-03
Completed LinkedList2

alekhya-12:
- Solved all problems on LeetCode, except 'Delete without head node'
- Need help in understanding how to update the prev pointer in 'Delete without head node' problem, as I am not sure, how the traversing is done when the head isn't given.
- Please review! Thanks!

### DIFF
--- a/BSTIterator.java
+++ b/BSTIterator.java
@@ -1,0 +1,61 @@
+/**
+Problem : BST Iterator
+Did it run on LeetCode: Yes
+
+Time Complexity : O(n), where n is the number of nodes
+Space Complexity : O(h), where h is the height of the tree
+
+Approach:
+1. The next element in the BST Iterator, can be derived from the inorder traversal of the tree.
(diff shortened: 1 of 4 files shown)


BSTIterator.java
```diff
@@ -0,0 +1,61 @@
+/**
+Problem : BST Iterator
+Did it run on LeetCode: Yes
+
+Time Complexity : O(n), where n is the number of nodes
```
samarthjj, line 5: O(1), hasNext. Amortized O(n), next().

samarthjj:
Good job Alekhya, Please review the inline comments!

samarthjj:
Good job Alekhya, Please review the inline comments!

samarthjj:
Good job Alekhya, Please review the inline comments!
For the **Delete without head node** there isn't a prev pointer to update, only val & next.

samarthjj:
Good job Alekhya, Please review the inline comments!
For the **Delete without head node** there isn't a prev pointer to update, only val & next.

PR #175, opened 2020-08-03
Completed. Please Review it !

sneharnair:


Problem-4_Intersection_of_list.py
```diff
@@ -0,0 +1,98 @@
+# APPROACH  1: Intermediate solution 
+# Time Complexity : O(max(a, b)), a: length of linkeed list A and b: length of the linked list B
+# Space Complexity : O(max(a, b))
```
samarthjj, line 3: The SC should be O(a) as only elems from listA are added to the hashset!

samarthjj:
Good job Sneha, Please see the inline comments!

samarthjj:
Good job Sneha, Please see the inline comments!

PR #234, opened 2020-10-11
PR LL2

jeevy222:


ramya-sruthi-d (commented):
Good job with code. Think about the time complexity of get intersection, what if the intersection is at the end of the list?

ramya-sruthi-d:
Good job with code. Think about the time complexity of get intersection, what if the intersection is at the end of the list?

PR #260, opened 2020-11-10
Linked list 2 completed - Please Review

GayatriHungund81194:


bstIterator.py
```diff
@@ -0,0 +1,34 @@
+# As taught in class, the iterator has concept of returning the element value only if the next elemnt exist so, I used dfs to traverse the tree and resturn the element only if requsted.
+#Time Complexity: O(h)
+#Space Complexity: O(1)
```
samarthjj, line 3: Time: Amortized O(1). O(h) happens once and gives O(1) TC for next h calls.
Space: O(h), space for the stack.

samarthjj:
Good work Gayatri, Please review the inline comments!
Please also finish the Intersection Lists problem as well.

samarthjj:
Good work Gayatri, Please review the inline comments!
Please also finish the Intersection Lists problem as well.

bstIterator.py
GayatriHungund81194: Will commit the list intersection problem

GayatriHungund81194: Added the missing problem, please review

PR #341, opened 2021-02-12
Completed Linked-List-2 : Please Review

Amrindersingh1:


BSTIterator.java
```diff
@@ -0,0 +1,43 @@
+// Time Complexity : O(n)
+// Space Complexity : O(1)
```
nsibal, line 2: TC would be O(1) in the average case and O(h) in the worst case.
SC would be O(h).

nsibal (commented):
Looks good, read inline comments!

nsibal:
Looks good, read inline comments!

PR #361, opened 2021-03-15
Completed LinkedList-2

SahanaVBhat:


binarySaerchIterator.java
```diff
@@ -0,0 +1,46 @@
+//Problem1 (https://leetcode.com/problems/binary-search-tree-iterator/)
+// Time Complexity : O(n)
+// Space Complexity : O(n)
```
nsibal, line 3: This TC is only for the constructor. The rest of your methods run in O(1) time. See if you can do better in terms of SC though.

intersectionofTwoLinkedList.java
```diff
@@ -0,0 +1,16 @@
+// Problem4  (https://leetcode.com/problems/intersection-of-two-linked-lists/)
+// Time Complexity : O(N*M)
+// Space Complexity : O(1)
```
nsibal, line 3: This solution has a ton of room for improvement. Can you try improving this?

nsibal (commented):
Looks good, read inline comments.

nsibal:
Looks good, read inline comments.

binarySaerchIterator.java
```diff
@@ -0,0 +1,46 @@
...
+public class binarySaerchIterator {
```
lyndseyBeil, line 4: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Class names should start with an upper case letter (NM_CLASS_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-class-names-should-start-with-an-upper-case-letter-nm-class-naming-convention).
Class names should be nouns, in mixed case with the first letter of each internal word capitalized. Try to keep your class names simple and descriptive. Use whole words-avoid acronyms and abbreviations (unless the abbreviation is much more widely used than the long form, such as URL or HTML).

PR #364, opened 2021-03-18
Completed Problem-1 : Please review!

mannatDev:


Problem-1.java
```diff
@@ -0,0 +1,55 @@
+// Time Complexity : height of the tree
+// Space Complexity : O(N)
```
nsibal, line 2: Amortized TC of your solution is O(1) and SC is O(h).

nsibal (commented):
Read inline comments.

nsibal:
Read inline comments.

PR #460, opened 2021-07-06
Added solutions for linked list 2

pandas5308:


Rooppesh:
Good Job with all the problems! Please add details such as TC, SC and approach!

Rooppesh:
Good Job with all the problems! Please add details such as TC, SC and approach!

PR #506, opened 2021-09-12
Completed Sprint Linked-List-2: Please review

tanviekirane:


Stnagam:
Good job! Please add TC SC as well

Stnagam:
Good job! Please add TC SC as well

PR #596, opened 2022-01-16
completed

gunjalgupta:


sonalgupta276 (commented):
Good job! Mention TC and SC for delete node

sonalgupta276:
Good job! Mention TC and SC for delete node

PR #759, opened 2022-07-20
Finished 2,3,4 Problems. Please Review. Thanks

VipinMamidi:


IntersectionOfTwoLinkedLists.java
```diff
@@ -0,0 +1,55 @@
+// Time Complexity : O(n)
```
jeelpatel18, line 1: m+n

jeelpatel18 (commented):
Good job. Submit 1st problem also when you solve it

jeelpatel18:
Good job. Submit 1st problem also when you solve it

PR #791, opened 2022-08-08
chore: python optimal solution

BitLegion1996:


ishika2741 (commented):
Good work! Kindly include tc and sc for delete without head ptr question as well.

ishika2741:
Good work! Kindly include tc and sc for delete without head ptr question as well.

PR #804, opened 2022-09-07
Completed. Please review 

Thirunaa:


Delete without Head pointer.py
```diff
@@ -0,0 +1,13 @@
+# TC : O(n)
```
rajchoksi1997, line 1: Time complexity can be optimized

rajchoksi1997 (approved):
Good Job

rajchoksi1997:
Good Job

PR #858, opened 2022-11-03
Linked list-2, please review and provide feedback

chaithanya2sai:


ReorderList.java
```diff
@@ -0,0 +1,58 @@
+package s30.linkedList.linkedList2;
+
+
+import s30.linkedList.ListNode;
+
+import java.util.Stack;
+
+
+// Time Complexity O(n)
+// Space complexity O(1)
```
daredevilteja, line 10: Please recheck SC, here an extra stack is used

daredevilteja:
Good Work! Keep it up.

PR #1007, opened 2023-09-26
completed intersection of 2lists, reorder list

krushikagujarati:


pvk3081 (commented):
2 problems are missing

pvk3081:
2 problems are missing

PR #1015, opened 2023-09-29
Completed Linked List-2: Please Review

kumarshrivatsav:


reorder_list.py
```diff
@@ -0,0 +1,55 @@
+# Time Complexity: O(n+m) where n and m is the number of nodes
```
ishika2741, line 1: There is just one list here , so how would you define n and m here

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1025, opened 2023-10-21
added all four solutions

merinjputhiyedath:


Nandha2Tech (commented):
Great Work! TC for reorder list would be O(n).

Nandha2Tech:
Great Work! TC for reorder list would be O(n).

PR #1090, opened 2024-02-22
Done with LinkedList-2

Prinitha:


IntersectionOfTwoLinkedLists.py
```diff
@@ -0,0 +1,18 @@
+# Definition for singly-linked list.
+from typing import Optional
```
ishika2741, line 6: Kindly add time and space as well

ishika2741 (commented):
Well done!

ishika2741:
Well done!

IntersectionOfTwoLinkedLists.py
Prinitha: Totally missed! Thanks for pointing it out!
TC: O(n+m) n and m being length of both the individual lists. Here we will take the max of them!
SC: O(1)

PR #1348, opened 2025-07-16
Completed intersectingLinkedList and reorderList problems

harsh23raval:


super30admin:
Strengths:
- The student has correctly implemented the solutions for both problems.
- The use of two pointers in the reorderList solution is efficient and correctly implemented.

Areas for Improvement:
- In the intersectingLinkedList solution, the final `if` condition is unnecessary because the loop ensures that `currB` and `currA` are equal when the loop exits.
- In the reorderList solution, the variable name `node` could be more descriptive, such as `prev` or `reversedHead`, to better indicate its purpose.
- The comments could be more detailed to explain why certain steps are taken, especially in the reorderList solution where the merging of the two halves might not be immediately clear to a reader.

PR #1351, opened 2025-07-19
Solved 3 problems

sethu95:


super30admin:
- problem1.java: The use of a static stack variable is incorrect as it will be shared across all instances of BSTIterator. This should be an instance variable. The dfs method can be optimized to use iteration instead of recursion to avoid potential stack overflow for very deep trees.
- problem2.java: The solution is well-implemented and follows the optimal approach. No major improvements are needed.
- problem3.java: The solution is correct and efficient. The commented-out alternative solution using a HashSet is also correct but less space-efficient, which is a good demonstration of understanding trade-offs.

Overall, the student demonstrates a good understanding of the problems and their solutions. The main area for improvement is ensuring that variables are correctly scoped (instance vs. static) and considering iterative approaches for recursion-heavy solutions to avoid stack overflow.

PR #1354, opened 2025-07-20
Done Linked-List-2

NUPUR0307:


super30admin:
Strengths:
- The student demonstrates a clear understanding of the problem and the optimal approach to solve it.
- The student correctly analyzes the time and space complexity.

Areas for Improvement:
- The student includes an unnecessary instance variable `root` in the BSTIterator class, which is not used after the initial constructor call. This could be removed to save memory.
- The method `dfs` could be made private since it is only used internally within the class.

PR #1363, opened 2025-08-05
Completed Linked-List-2

joelkny97:


super30admin:
- **BSTIterator (Exercise_1.cpp)**: The solution is excellent. No improvements needed.
- **Reorder List (Exercise_2.cpp)**: The solution is well-implemented. One minor suggestion is to handle edge cases like an empty list or a single-node list explicitly, though the current implementation handles them implicitly.
- **Delete Node (Exercise_3.cpp)**: The solution is correct and efficient. However, it might be worth mentioning that this approach doesn't work for deleting the last node in the list, which is a limitation of the problem itself.
- **Intersection of Two Linked Lists (Exercise_4.cpp)**: The solution is correct, but the condition `if (currA == nullptr && currB == nullptr) return nullptr;` is redundant because the loop will terminate when both pointers are `nullptr` and return `nullptr` anyway. Removing this line would simplify the code without affecting correctness.

Overall, the student has demonstrated a strong understanding of the problems and provided efficient, well-structured solutions. The minor suggestions are for further refinement rather than corrections.

PR #1365, opened 2025-08-18
Janhvi Guha Linked-List-2

Janhvi2000:


super30admin:
Strengths:
- The solutions are correct and handle edge cases well.
- The code is clean, readable, and well-commented.
- Time and space complexities are correctly analyzed.

Areas for Improvement:
- For ex1.py (BSTIterator), consider using a stack to perform in-order traversal incrementally. This would reduce space complexity from O(n) to O(h) and make the solution more efficient for large trees.
- For ex2.py (reorderList), the logic is correct, but you could add a comment explaining why the loop condition in the merge step works (i.e., why you don't need to check both `slow` and `fast`).

PR #1366, opened 2025-08-18
completed Linked-List-2

nitishred13:


super30admin:
Strengths:
- The implementation of BSTIterator is correct and follows the same approach as the reference solution.
- The code is clean, well-organized, and easy to understand.
- Proper use of Stack and helper methods.

Areas for Improvement:
- The time complexity comment should be corrected to "amortized O(1)" instead of O(logn).
- The space complexity comment should be O(h) (height of tree) rather than O(logn), as logn only applies to balanced trees.
- The other files included in the submission (DeleteNode.java, IntersectionLL.java, ReorderList.java) are not relevant to the BSTIterator problem and should be submitted separately if they are solutions to other problems.

PR #1370, opened 2025-08-18
Linked-List-2 Completed

arpitha-nagaraj9791:


super30admin:
Strengths:
- The solutions are correct and efficient, matching the reference solutions in terms of time and space complexity.
- The code is well-organized, with clear separation of concerns and helper methods where appropriate.
- Comments are helpful and accurately describe the approach.

Areas for Improvement:
- In the IntersectionOf2LinkedList solution, the condition `if(currA == null && currB == null) return null;` is redundant because the loop will naturally terminate when both pointers are null. This check can be removed without affecting correctness.
- The class name for BSTIterator should be `BSTIterator` to match the problem requirements on LeetCode, though this might be a minor issue depending on the context.

PR #1375, opened 2025-08-22
Completed Linked-List-2

AneerudhS27:


super30admin:
Strengths:
- The student demonstrates a strong understanding of data structures and algorithms.
- The solutions are well-implemented and follow best practices.
- The use of comments to explain the logic is commendable.

Areas for Improvement:
- In intersectionOf2LL.cpp, the condition `if(currA ==nullptr && currB == nullptr) return nullptr;` is redundant because the loop condition `while(currA != currB)` will naturally handle this case. The loop will terminate when both pointers reach nullptr simultaneously.
- In reorderList.cpp, the variable names `slow` and `fast` are reused after finding the middle, which could be confusing. Consider using different names for clarity.
- The inclusion of unnecessary headers like `<bits/stdc++.h>` should be avoided in favor of specific headers needed for the solution.

PR #1383, opened 2025-09-11
Implemented Linked-List-2

johar-inderpreet:
Implemented Linked-List-2


super30admin:
Strengths:
1. Correct implementation of the BSTIterator with lazy evaluation.
2. Clear and concise code with good method naming.
3. Includes a main method for testing, which is helpful for verifying correctness.

Areas for Improvement:
1. The comments in the BSTIterator.java file mention "Time Complexity: O(h)" and "Space Complexity: Amortized O(1)", which could be more precise. The time complexity for next() is amortized O(1), and the space complexity is O(h).
2. The main method includes a test case that attempts to call next() when the stack is empty, which will throw an EmptyStackException. It would be better to handle this case gracefully or document it clearly.

PR #1398, opened 2025-10-03
Linked-List-2

mdashora:


super30admin:
- The main issue is that you solved the wrong problem. The problem was about implementing a BSTIterator for a binary search tree, not reordering a linked list. Please carefully read the problem statement before starting to code.
- Your linked list reordering solution is technically sound and well-implemented, showing good understanding of linked list manipulation techniques (finding middle, reversing, merging).
- For the actual BSTIterator problem, you would need to implement an iterator that allows in-order traversal of a BST using a stack, with next() and hasNext() operations.
- Pay attention to problem requirements and make sure your solution matches what's being asked.

PR #1422, opened 2026-02-04
Linked-List-2

amitmittal117:


super30admin:
Strengths:
- Your solution is correct and straightforward to understand.
- The code is clean and well-commented.

Areas for Improvement:
- The problem's follow-up asks for O(h) memory usage, but your solution uses O(n) space. Consider using an iterative in-order traversal with a stack to achieve O(h) memory.
- For large trees, the recursive in-order traversal might cause recursion depth issues. An iterative approach would be more robust.
- The reference solution uses a stack to simulate the in-order traversal step-by-step, which allows for lazy evaluation (only traversing as needed) and better memory efficiency.

Suggested Improvement:
Instead of storing all values upfront, you can use a stack to keep track of the nodes. During initialization, push all left nodes until the smallest element. For `next()`, pop the top node, push all left nodes of its right child, and return the value. This way, the stack never holds more than O(h) nodes.

Example code for the improved approach:
```python
class BSTIterator:
    def __init__(self, root):
        self.stack = []
        self._push_left(root)
    
    def _push_left(self, node):
        while node:
            self.stack.append(node)
            node = node.left

    def next(self) -> int:
        node = self.stack.pop()
        self._push_left(node.right)
        return node.val

    def hasNext(self) -> bool:
        return len(self.stack) > 0
```

This approach meets the O(h) space requirement and has amortized O(1) time for `next()`.

super30admin:
Strengths:
- The BST Iterator solution is straightforward and easy to understand.
- The in-order traversal is correctly implemented.
- The `next()` and `hasNext()` methods are efficient (O(1) time) for each call.

Areas for Improvement:
- The space complexity for the BST Iterator is O(n), which is not optimal for large trees. The problem requires O(h) space, which can be achieved using an iterative approach with a stack (as in the reference solution). This is important because the problem constraints and follow-up specifically ask for O(h) memory.
- Consider using an iterative in-order traversal that only stores the left path of the tree until needed. This would reduce the space to O(h).

For the other problems:
- The solution for deleting a node without a head pointer is correct and efficient.
- The solution for finding the intersection of two linked lists is correct and efficient.

PR #1426, opened 2026-02-14
solved

jeetusingh247:


super30admin:
Your solution for BSTIterator is excellent. It correctly implements the iterator with the required time and space complexity. The code is clean and easy to understand. However, note that the problem only asks for the BSTIterator implementation. Including solutions for other problems in the same file might be confusing and is not necessary. In the future, ensure that you only submit the code relevant to the problem at hand.

One minor suggestion: the dfs method could be renamed to something more descriptive, like pushAllLeft, to make it clear that it pushes all left children onto the stack. This is a common naming convention in such solutions.

Overall, great job on the BSTIterator implementation.

super30admin:
No specific feedback provided. Refer to GitHub Comment.

PR #1429, opened 2026-03-01
Completed LinkedLists-2

sarvanibaru:


super30admin:
Your solution for the BSTIterator problem is excellent. It correctly implements the in-order traversal using a stack with lazy evaluation, which meets the problem requirements. The time and space complexities are correctly analyzed, though note that the space complexity is more precisely O(h) (the height of the tree) rather than O(n) in the worst case. However, in the worst case (a skewed tree), h = n, so O(n) is not wrong. But for the purpose of this problem, O(h) is the expected answer as per the follow-up.

The code is clean and easy to understand. The helper method is well-named and efficiently handles pushing the left nodes. One minor point: the helper method is called `helper`, but it might be more descriptive to name it `dfs` or `pushLeft` to indicate its purpose more clearly. However, this is a matter of preference.

Overall, great job! You have successfully implemented the BST iterator with the required time and space complexities.

PR #1430, opened 2026-03-01
[reorderList && DeleteNode && BSTIterator && GetIntersection]

takchiks:


super30admin:
Your solution for BSTIterator is correct and efficient. It correctly uses a stack to implement the in-order traversal iterator. The time and space complexities meet the requirements. However, consider using an iterative approach for the DFS (pushing left nodes) instead of recursion to avoid potential stack overflow for very deep trees. For example, in the helper function, you can use a while loop to traverse left and push nodes without recursion. This would make the solution more robust and avoid recursion overhead.

Also, note that you included code for other problems in the same file. It's important to submit only the code for the problem being evaluated. For the BST Iterator problem, your solution is good, but the recursive DFS could be improved to iterative.

Here is a slight modification to your helper function to make it iterative:
```
private void helper(TreeNode root) {
    while (root != null) {
        st.push(root);
        root = root.left;
    }
}
```
This iterative version does the same thing without recursion.

Overall, your solution is correct and efficient, but with the iterative DFS it would be even better.

PR #1432, opened 2026-03-05
Adding all the solutions

subbu4061:


super30admin:
Your solution for the BST Iterator problem is correct and efficient. You have implemented the iterative in-order traversal using a stack, which is the standard approach. The time and space complexities are optimal. Well done!

However, note that the problem requires the implementation of a BST Iterator, and you have provided solutions for other problems as well. In the future, make sure to only submit the relevant code for the problem at hand. Also, pay attention to file naming to avoid confusion.

For the BST Iterator solution:
- The code is clean and follows good practices.
- The `dfs` helper function is appropriately named and used.
- The stack is initialized correctly in the constructor.

One minor suggestion: in the `dfs` method, you use `st.add(root)`, which is equivalent to `st.push(root)` since `Stack` in Java uses `add` as well, but it's more conventional to use `push` for stacks to make the intent clear. However, both work.

PR #1438, opened 2026-05-02
Completed Linked List 2

pratikb0501:


super30admin:
### BST Iterator (Problem_1.py)
Your solution is on the right track by using a stack to simulate the in-order traversal. However, there are a few issues that need to be addressed:

1. **Recursion Depth**: The recursive implementation of `_dfs` may lead to a stack overflow for large trees (with up to 10^5 nodes) because Python has a limited recursion depth. Instead, you should use an iterative approach to traverse the left subtree. This is what the reference solution does: it uses a while loop to push all left nodes without recursion.

2. **Initialization**: In the constructor, you push the root and then call `_dfs(root.left)`. This is correct in terms of building the stack, but it relies on recursion. The reference solution initializes the stack and then calls a helper function (which is iterative) to push the root and all left children. You can do the same.

3. **Next Method**: Your `next` method is almost correct: it pops the top node, then calls `_dfs` on the right child. However, because `_dfs` is recursive, it might cause recursion depth issues. You should replace the recursive `_dfs` with an iterative one.

Here's how you can modify your code to be iterative:

- Change the `_dfs` function to be iterative. Actually, you can integrate the iterative left traversal into the constructor and the `next` method. The reference solution does this by having a helper function that uses a while loop.

Revised code example (based on the reference solution):

```python
class BSTIterator:

    def __init__(self, root):
        self.st = []
        self._push_all_left(root)

    def _push_all_left(self, node):
        while node:
            self.st.append(node)
            node = node.left

    def next(self):
        top = self.st.pop()
        self._push_all_left(top.right)
        return top.val

    def hasNext(self):
        return len(self.st) > 0
```

This iterative approach avoids recursion and is efficient in both time and space. It also handles large trees without recursion depth errors.

4. **Correctness**: Your current implementation might work for small trees, but for large trees it might fail due to recursion limits. The revised iterative approach is more robust.

Overall, you have the right idea, but you need to avoid recursion to make the solution scalable.

VERDICT: NEEDS_IMPROVEMENT

---

### Reordering of Linked List (Problem_2.py)
Your solution correctly identifies the approach of finding the middle, reversing the second half, and merging. The code is clean and easy to follow. However, there is a critical issue with the space complexity due to the recursive reversal of the linked list. For large lists (with up to 50,000 nodes), the recursion depth may exceed the limit and cause a stack overflow. 

To improve:
- Replace the recursive reverse function with an iterative one. This will reduce the space complexity from O(n) to O(1), which is necessary for large inputs.
- Alternatively, you can keep the recursive reverse if the problem constraints are small, but here they are not.

Here is how you can implement iterative reverse in Python:
```python
def reverseLL(self, head):
    prev = None
    curr = head
    while curr:
        next_temp = curr.next
        curr.next = prev
        prev = curr
        curr = next_temp
    return prev
```

Also, note that your current solution returns `head` in the `reorderList` function. However, the problem does not require returning the head (it modifies in-place). But in Python, since we are modifying the list in-place and the function does not need to return anything (the problem expects the list to be modified), you can avoid returning. However, returning the head is not wrong, but it is unnecessary.

Another minor point: the base case in `reverseLL` (if not head or not head.next) is correct, but the iterative reverse can handle that without a base case.

Overall, the logic is correct, but the recursive reverse is a significant drawback.

VERDICT: NEEDS_IMPROVEMENT

---

### Deletion of Node (Problem_3.py)
Your solution is on the right track and works for non-tail nodes by copying the data from the next node and then skipping it. However, you need to consider the edge case when the node to be deleted is the last node. In that case, since you don't have access to the head, you cannot delete it (as you need to update the previous node's next pointer). The problem constraints state that you are given a reference to the node to be deleted, but if it's the last node, you should not do anything (or the problem might assume that the node is not the last? Actually, the problem says: "you are given the reference/pointer to the node which is to be deleted", and it does not guarantee that it is not the tail. The example inputs do not include deleting the tail, but the constraints say N>=1, and the node to be deleted is one of the nodes.

Looking at the problem statement: "you are not given the head node ... but you are given the reference/pointer to the node which is to be deleted". The problem does not specify that the node is not the tail. However, the reference solution in Java handles the case by checking if the node is null or the next is null. In Python, you should do the same.

You should add a check to see if the node is the last node. If it is, you cannot delete it (because without the head, you cannot update the previous node's next to null). But note: the problem says "you have to delete that node only with this much of information". So if the node is the last node, it might be an invalid input? Actually, the problem constraints say the node to be deleted is given, and it is one of the nodes. So it could be the last.

However, the problem's example output does not show deleting the last node. But the problem says: "The last line contains the node to be deleted." So the input value is the data of the node to be deleted. It might be the last node.

But wait: the driver code is responsible for printing the list after deletion. If the node to be deleted is the last node, then we cannot delete it without the head. So the problem likely assumes that the node to be deleted is not the tail? Actually, the problem says: "you are not given the head node", meaning that you cannot traverse to find the previous node. So the problem expects the same approach as the reference solution: which works only for non-tail nodes.

But the problem constraints do not say that the node is not the tail. So we must handle it.

In the reference solution (Java), if the node is the tail, it does nothing. But that leaves the tail node still there? Actually, that is incorrect because the problem says to delete the node. However, without the head, it is impossible to delete the tail. So the problem likely guarantees that the node is not the tail.

Looking at the problem statement: "you are given the reference/pointer to the node which is to be deleted". This implies that the node is not the tail because if it were, we couldn't delete it. So the problem assumes that the node is not the tail.

But to be safe, the reference solution in Java checks for tail and returns. So we should do the same.

Your code should check if the given node is the last node. If it is, you cannot delete it. So you should return without doing anything.

In Python, you can do:

if del_node is None or del_node.next is None:
    return

Then proceed as you did.

So your solution is almost correct, but missing this check.

Also, in Python, it's better to use snake_case for variable names. So `nextNode` should be `next_node`.

VERDICT: NEEDS_IMPROVEMENT

---

###  Intersection of Two Lists (Problem_4.py)
Your solution is on the right track by using the two-pointer technique, which is efficient in both time and space. However, there are a few issues that need to be addressed:

1. **Redundant Condition**: The inner condition `if currA == currB:` inside the while loop is unnecessary because the while loop already ensures that `currA != currB`. You can remove this condition.

2. **Handling No Intersection**: The main problem is that your code does not handle the case when there is no intersection correctly. In the reference solution, after moving both pointers, it checks if both are null and returns null. Your code lacks this check, which can lead to an infinite loop when the lists do not intersect.

3. **Pointer Movement**: In the reference solution, both pointers are moved first, and then checks are performed. Your code moves each pointer independently and resets them immediately when they become null. This is not exactly the same. The correct approach is to move both pointers and then check if they are equal (which includes both being null).

To fix your code, you should:
- Remove the inner `if currA == currB` condition.
- Move both pointers first, then check if they are equal. But actually, the while condition checks for equality at the start of each iteration.
- Alternatively, you can structure your code exactly like the reference solution: 
   - Inside the loop, first move both pointers to their next.
   - Then, check if both are null (return null).
   - Then, if one is null, reset it to the head of the other list.

Here is how you can correct your code:

```python
class Solution:
    def getIntersectionNode(self, headA, headB):
        currA, currB = headA, headB
        while currA != currB:
            currA = currA.next if currA else headB
            currB = currB.next if currB else headA
        return currA
```

Wait, actually the above corrected code is common in Python solutions. It works because:
- If there is an intersection, the pointers will meet at the intersection node.
- If there is no intersection, both pointers will eventually become null at the same time (after traversing m+n nodes), and then the while loop will break (since null == null) and return null.

So the corrected code is very concise. The student's code almost had it, but the redundant condition and the separate handling of the pointer movement caused the issue.

Another way to write the same (without the redundant condition) is:

```python
class Solution:
    def getIntersectionNode(self, headA, headB):
        a, b = headA, headB
        while a != b:
            a = a.next if a else headB
            b = b.next if b else headA
        return a
```

This is the standard solution for Python.

So the student's code can be fixed by removing the inner condition and simplifying the movement.

VERDICT: NEEDS_IMPROVEMENT

PR #1439, opened 2026-05-07
Complete Linked-List-2

dhruvil15:


super30admin:
### BST Iterator (Problem52.java)
Your solution correctly implements the BST iterator with in-order traversal. However, using recursion in the `dfs` method can be problematic for large trees because it may cause a stack overflow. Consider using an iterative while loop instead, as shown in the reference solution. This change will make your code more efficient and robust. Here's how you can modify your `dfs` method:

```java
private void dfs(TreeNode root) {
    while (root != null) {
        this.stack.push(root);
        root = root.left;
    }
}
```

This iterative approach avoids recursion and uses only the stack data structure without additional call stack overhead. Keep up the good work!

VERDICT: NEEDS_IMPROVEMENT

---

### Reordering of Linked List (Problem53.java)
Strengths:
- The solution correctly implements the required algorithm.
- The code is clean and easy to follow.
- The reverse function is implemented efficiently.
- The merging step is handled correctly by updating pointers in the right order.

Areas for Improvement:
- While the code is correct, it is important to handle edge cases explicitly. For example, if the list is empty or has only one node, the code should still work. In this case, the code should be tested with such inputs. However, given the constraints (number of nodes in the range [1, 5*10^4]), the head will never be null, but it's good practice to check for null in general.
- The variable names are consistent, but in the reverse function, the variable 'temp' is used to store the next node. This is acceptable, but some developers prefer more descriptive names like 'nextNode' to avoid confusion with other temporary variables.
- The comments are helpful, but they could be expanded to explain the approach briefly.

VERDICT: PASS

---

### Deletion of Node (Problem54.java)
Your solution is excellent and matches the expected approach perfectly. You have correctly identified the time and space complexity. The code is clean and efficient. 

One minor consideration: the problem states that the node to be deleted is guaranteed to not be the tail node? Actually, the problem says "you are given the reference/pointer to the node which is to be deleted", but it doesn't explicitly say it won't be the tail. However, the constraints in the problem might imply that the node to be deleted is always valid and not the tail? But in your solution, you have a check for `del.next == null` which handles the case when the node is the tail. However, the problem says "you have to delete that node only with this much of information", and if it's the tail, we cannot delete it without the head. So the problem likely ensures that the node is not the tail. But your check is good practice.

Overall, great job!

VERDICT: PASS

---

###  Intersection of Two Lists (Problem55.java)
Your solution shows a good understanding of the two-pointer technique, which is a common approach for this problem. However, there are a few issues that need to be addressed:

1. **Infinite Loop Risk**: The current code may run into an infinite loop when the lists do not intersect. This is because the condition to return null (`if (currA == null && currB == null) return null;`) is checked only after moving both pointers. But when the lists are of different lengths and non-intersecting, the pointers will eventually both become null at the same time only if the lists are of the same length. For lists of different lengths, the pointers will reset and traverse the other list, but they might not both become null simultaneously again. The reference solution avoids this by ensuring that after both pointers have traversed both lists, if there is no intersection, they will both become null at the same time and exit the loop.

2. **Null Pointer Exception**: If either `headA` or `headB` is null, then `currA.next` or `currB.next` will throw a NullPointerException. You should handle the case where either list is null at the start.

3. **Code Structure**: The code is concise and readable, which is good. However, the logic needs to be adjusted to correctly handle the traversal.

To fix the issue, you can adjust the order of operations. The reference solution first checks if the current pointers are equal (which is the loop condition). Then, it moves both pointers. If either pointer becomes null, it resets to the head of the other list. This ensures that both pointers traverse exactly m + n nodes if there is no intersection, and then both become null simultaneously, breaking the loop.

Here is a corrected version of your code in Java (similar to the reference solution in C++):
```java
public class Solution {
    public ListNode getIntersectionNode(ListNode headA, ListNode headB) {
        if (headA == null || headB == null) return null;
        ListNode currA = headA;
        ListNode currB = headB;
        while (currA != currB) {
            currA = (currA == null) ? headB : currA.next;
            currB = (currB == null) ? headA : currB.next;
        }
        return currA;
    }
}
```

This version avoids the infinite loop by resetting the pointers only when they become null and then continuing. After both pointers have traversed both lists, they will both be null at the same time if there is no intersection, and the loop will break.

Keep up the good work! Your initial approach was close to the optimal solution, but just needed a slight adjustment in the pointer reset logic.

VERDICT: NEEDS_IMPROVEMENT

